PR #70, opened 2026-08-29
fix(build): unbreak master — SUPERSEDED BY #66

srpatcha:
**P0.** `master` cannot run. Reproduced from a clean checkout:

```
    else:
    ^^^^
SyntaxError: invalid syntax

$ pytest
2 errors during collection
```

Four defects. Every one reached master because the file it lives in is imported **lazily**, so nothing on the merge path ever executed it.

| # | Defect | Symptom |
|---|---|---|
| 1 | Two consecutive `else:` blocks in `configure()` | SyntaxError; `tests/unit` uncollectable |
| 2 | `"ninja"` listed among backends needing no configure | `configure("ninja")` silently succeeds — the exact regression `test_dispatch.py` guards against |
| 3 | `NinjaBackend._object_path` called twice, defined nowhere | `AttributeError` on **every** ninja build |
| 4 | ninja invoked as `python -m ninja` | "No module named ninja" on a machine with ninja installed |

Kept the RuntimeError arm of #1 because `tests/unit/test_dispatch.py` documents the intent — an unhandled backend must fail loudly rather than do nothing and let the caller report a false "Build completed successfully".

Also unified the exception type: `build()` and `clean()` raised `ValueError` for the same condition `configure()` reports as `RuntimeError`, so no single `except` guarded the dispatcher.

**Shared libraries:** the `link_shared` rule was declared and never used — edges went through the generic `link` rule with `-shared` pushed into ldflags. Functional, but it left the rule dead and broke `test_shared_library_uses_shared_link_rule`. Edges use `link_shared` now, with the platform flag in the rule so darwin's `-dynamiclib` is decided in one place.

## Verification

| | before | after |
|---|---|---|
| `pytest` | **2 errors during collection** | **202 passed** |
| `ebuild build` | Python traceback | reaches the compiler |

`ebuild build` now reports a real compile error instead of a traceback. The remaining first-run gap — templates including `<eos/hal.h>` with no path — is #64's scope, not this fix's.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


srpatcha:
## Superseded by #66 — take that one instead

I wrote this before seeing #66. Both repair the same `SyntaxError` in `ebuild/build/dispatch.py` that stops `master` collecting its own test suite. **#66 came first, and it is the change I have approved.**

Verified on `origin/master`: #66 alone takes the suite from `2 errors in 0.25s` to **206 passed**.

Comparing the two, #66 covers three of the four defects I found. The one it does not:

```python
ebuild/build/dispatch.py:244   [sys.executable, "-m", "ninja", "-C", ...]
ebuild/cli/commands.py:690     [sys.executable, "-m", "ninja", "-f", ...]
```

`commands.py:690` is the build path. It works today only because `pyproject.toml` declares `ninja>=1.11`, which puts the PyPI wheel in the environment — but a machine with a perfectly good `ninja` on PATH is ignored, and the moment that wheel is absent the build fails with `No module named ninja` rather than anything a developer can act on.

I have noted that on #66 as a follow-up. It is three lines and does not belong in a hotfix:

```python
def ninja_command() -> list:
    exe = shutil.which("ninja")
    return [exe] if exe else [sys.executable, "-m", "ninja"]
```

Leaving this open rather than closing it, since that is your call — but nothing here needs to merge, and merging both would conflict.

The three feature PRs that were branched on top of this one (#71 footprint, #72 doctor, #74 add/summary) are **rebased onto plain `master`** now. None of them touches `dispatch.py`, so they no longer depend on this branch or on #66, and all three report MERGEABLE.

srpatcha:
Superseded by #66, which has merged. `master` now parses and the suite runs — 287 passed, 1 skipped from a clean clone. Nothing here is still needed.

PR #75, opened 2026-08-29
Validate toolchain extra_cflags/extra_ldflags as string lists

swayam-2003:
## Summary
- Remove duplicate `else` in `BackendDispatcher.configure()` that caused a SyntaxError and blocked importing the package.
- Unknown backends now raise `RuntimeError` consistently in `configure()`/`build()`.
- Validate `toolchain.extra_cflags` and `toolchain.extra_ldflags` as string lists so scalar YAML values cannot silently split into per-character flags.

## Approach
Followed existing target-field validation patterns in `config.py` and regression tests in `tests/unit/test_dispatch.py`.

## Testing
```powershell
python -m pip install -e ".[dev]"
python -m pytest tests/unit/test_dispatch.py tests/ebuild/test_config_validation.py -v
```
21 tests passed locally.

## DCO
Signed-off-by: Swayam Nayak <154440440+swayam-2003@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Fixed
+- **`load_config`:** `toolchain.extra_cflags` and `toolchain.extra_ldflags` must be YAML lists of strings so scalar values cannot silently split into per-character compiler flags.
 - **Ninja backend: header changes now trigger a rebuild.** The generated `cc`
   rule declared no depfile, so Ninja only knew about the sources listed in
   `build.yaml`. Editing a header left stale object files in place and the build
--- a/ebuild/core/config.py
+++ b/ebuild/core/config.py
@@ -165,13 +165,29 @@ def _parse_target(raw: Any) -> TargetConfig:
 
 def _parse_toolchain(raw: Dict[str, Any]) -> ToolchainConfig:
     """Parse toolchain section into a ToolchainConfig."""
+    extra_cflags = raw.get("extra_cflags", [])
+    extra_ldflags = raw.get("extra_ldflags", [])
+
+    for field_name, value in (
+        ("extra_cflags", extra_cflags),
+        ("extra_ldflags", extra_ldflags),
+    ):
+        if not isinstance(value, list):
+            raise ConfigError(
+                f"Toolchain field '{field_name}' must be a list."
+            )
+        if not all(isinstance(item, str) for item in value):
+            raise ConfigError(
+                f"Toolchain field '{field_name}' must contain only strings."
+            )
+
     return ToolchainConfig(
         compiler=raw.get("compiler", "gcc"),
         arch=raw.get("arch", "x86_64"),
         prefix=raw.get("prefix"),
         sysroot=raw.get("sysroot"),
-        extra_cflags=raw.get("extra_cflags", []),
-        extra_ldflags=raw.get("extra_ldflags", []),
+        extra_cflags=extra_cflags,
+        extra_ldflags=extra_ldflags,
     )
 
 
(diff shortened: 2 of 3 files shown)


srpatcha:
## The flags validation is the valuable half, and I confirmed it

`extra_cflags: -O2 -Wall` is the natural thing to write, and YAML gives you a string. `ToolchainConfig` annotates the field `List[str]` and nothing enforces it, so the string is passed straight through to code that iterates it:

```
YAML gives:        '-O2 -Wall'  (str)
iterating it:      ['-', 'O', '2', ' ', '-', 'W', 'a', 'l', ...]
```

Every character becomes a separate compiler flag. The build then fails with something about an unrecognised option `-`, which points at nothing the developer wrote. A type annotation is documentation, not a check, and this is the case that shows the difference.

Rejecting it at load with a message naming the field is right. It is the same treatment #73 gives a malformed `packages:` section, and the same class as several others in this repo: a value of the wrong shape silently accepted, and the consequence surfacing somewhere unrelated.

## The dispatcher half is superseded

**Three PRs fix the same `SyntaxError`: #66, #70 and this one.** I have approved #66 — it came first and is the wider change. My own #70 I have marked superseded for the same reason.

Standardising on `RuntimeError` for unknown backends is correct and is what #66 does too, so that part lands either way.

## What I would like

Rebase onto #66 and reduce this to the `config.py` validation plus its tests. That should be a small branch with no conflicts, and I will merge it — the flags bug is not fixed by #66 or by anything else open, and it is the piece that would otherwise be lost.

Worth extending while you are there, if you want: `extra_ldflags` and `extra_cflags` are not the only `List[str]` fields that a scalar would break. `sources`, `includes`, `defines` and `depends` on a target have the same shape, and `sources: src/main.c` is at least as easy to write as the list form.

## Verification

Confirmed the scalar-splitting behaviour on current `master` with a direct YAML load. Not verified this branch merged — it conflicts with #66 in `dispatch.py`, which is the overlap described above.

swayam-2003:
Thanks for the review @srpatcha — rebased onto latest master and removed the dispatch.py changes (superseded by #66). This PR now contains only the toolchain flag validation in config.py + tests + CHANGELOG. Ready for another look.

srpatcha (commented):
The toolchain-flag validation here is a genuine find and I want it kept. The
syntax repair in the title is not in the diff, though — I checked, because two
other PRs claim the same fix and I wanted to know which actually deliver it.

## The syntax error is still present on this branch

```
$ git show pr75:ebuild/build/dispatch.py > /tmp/d75.py
$ python3 -c "import ast; ast.parse(open('/tmp/d75.py').read())"
  line 133: invalid syntax
   129                 f"Unknown build backend '{backend}'. "
   130                 f"Supported backends: {', '.join(sorted(ALL_BACKENDS))}"
   131             )
   132
   133         else:
   134             raise RuntimeError(
```

Byte-identical to `master`. For comparison, #66 and #78 both parse after their
changes.

Tracked as #87. #66 is the one I would take for it — it removes the duplicated
error path that caused the splice rather than deleting the stray `else:`.

## The flags validation is the part worth saving

This is a real bug and I verified it independently. `extra_cflags` given as a
string is iterated character by character:

```python
>>> list('-O2 -Wall')
['-', 'O', '2', ' ', '-', 'W', 'a', 'l', 'l']
```

Every character becomes its own compiler argument. The build then fails with
something unrelated-looking, or worse, silently drops the flags the developer
asked for — a `-O2` that never reached the compiler is not a visible failure, it
is a slow binary nobody can explain.

Rejecting a scalar where a list is expected, with a message naming the field, is
exactly right, and I have not seen this covered anywhere else in the open PRs.

## Suggestion

Rebase onto #66 and drop the `dispatch.py` hunks that overlap it, leaving this
PR as the flags validation alone. It then reviews on its own merits, has no
conflict to settle, and the part that is uniquely yours is not blocked behind a
syntax fix that two other PRs are also trying to land.

Retitling would help too — as it stands the title promises a repair the diff does
not contain, which is how a reviewer ends up merging it expecting `master` to be
fixed.

Happy to approve once it is scoped to the validation.

swayam-2003:
Rebased onto latest upstream/master (d3958f7). Previous CI failures were caused by a stale base that still had the dispatch.py SyntaxError — that blocked pytest collection entirely. This PR now contains only toolchain flag validation + tests + CHANGELOG. All 42 related tests pass locally after rebase.

PR #76, opened 2026-08-29
fix(deps): resolve sibling eBoot as well as eboot

ShrenikMensinkai:
## Summary

Sibling repo lookup used only `../eboot`. The public repo and many local trees are named `eBoot`, so Linux CI and a case-sensitive workspace would not find the bootloader. This PR tries both names in the dependency manager and the pytest fixture.

This addresses:

- **Issue:** case-sensitive miss of the eBoot sibling
- **Approach:** alias list `eboot`, `eBoot` on the sibling step only
- **Testing:** host tests with a case-sensitive `is_dir` mock; 4 passed
- **Limits:** sibling step only; does not change cache, env, or CLI overrides

## Type of Change

- [ ] feat — New feature
- [x] fix — Bug fix
- [x] test — Add or fix tests

## Changes

- Resolve `"eboot"` from `../eboot` or `../eBoot`
- Point the eboot pytest fixture at whichever sibling exists
- Add tests for camel-case, lowercase, and missing siblings

## Testing

- [x] Unit tests pass (`python -m pytest tests/ebuild/test_deps_manager.py -v`)
- [ ] Integration tests pass

## Pre-Submission Checklist

- [ ] All existing tests pass
- [ ] Branch is rebased on latest master

## Related Issues

None. Found while reviewing workspace layout against GitHub’s `eBoot` repo name.

## Screenshots / Logs

```
4 passed in 0.07s
```

## Additional Notes

- Does not alter `ebuild/core/eboot` (deprecated).
- Windows already treated `eboot` and `eBoot` as the same folder; this is for Linux and CI.
- Integration CLI already had a local case fallback; the shared `DepsManager` path did not.

### DIFF
--- a/ebuild/deps/manager.py
+++ b/ebuild/deps/manager.py
@@ -14,7 +14,7 @@ import os
 import subprocess
 import warnings
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 import yaml
 
@@ -29,6 +29,12 @@ from ebuild.deps import (
 
 # Known repo names
 KNOWN_REPOS = ("eos", "eboot")
+
+# Sibling directory names. GitHub and local checkouts often use eBoot.
+SIBLING_DIR_NAMES: Dict[str, Tuple[str, ...]] = {
+    "eos": ("eos",),
+    "eboot": ("eboot", "eBoot"),
+}
 
 # Environment variable names for path overrides
 ENV_PATH_VARS = {
@@ -46,7 +52,7 @@ class DepsManager:
     2. Environment variables ``EBUILD_EOS_PATH`` / ``EBUILD_EBOOT_PATH``
     3. ``~/.ebuild/config.yaml`` custom ``path:`` override
     4. ``~/.ebuild/repos/<name>/`` (cached git clone)
-    5. Sibling directory ``../<name>/`` (workspace layout)
+    5. Sibling directory ``../<name>/`` (workspace layout; ``eboot`` also tries ``eBoot``)
     6. Embedded ``core/<name>/`` (legacy fallback — prints deprecation warning)
     """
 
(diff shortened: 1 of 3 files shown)


srpatcha (approved):
## Verified — and this is the third instance of the same bug in the org

`../eboot` does not match a directory called `eBoot` on a case-sensitive filesystem, so sibling resolution failed on Linux for the one repository whose canonical name is mixed-case. The clone the workspace actually contains is `eBoot`, because that is what GitHub serves.

I hit the identical bug in EoSim: `find_repos()` held a hardcoded list of lowercase names — `eai`, `eni`, `eipc`, `eboot`, `ebuild-tool` — and found **2 of the 19 repositories** in the workspace. And ebuild's own `ebuild setup` still clones into `~/.ebuild/repos/eboot` while the remote is `eBoot`, which is why the cache path and the sibling path disagree in the first place.

Three separate places, one cause: a repository name written down in a different case from the one the filesystem stores.

## On the approach

An explicit alias table is the right choice over case-insensitive matching. It is predictable, it is greppable, and it fails loudly if a new repository is added without an entry — where a `lower()` comparison would silently paper over the next mismatch and make the underlying inconsistency permanent.

Confining it to a module-level `SIBLING_DIR_NAMES` rather than sprinkling `or "eBoot"` at the call site is what makes it maintainable; the next person adding a sibling has one place to look.

Updating the docstring line — `sibling directory ../<name>/ (workspace layout; eboot also tries eBoot)` — matters more than it looks, because the resolution order is the thing a developer debugs against when a package is not found.

## One thing worth considering, not blocking

The canonical fix is upstream of this: the repository is `eBoot`, so the key could be `eBoot` with `eboot` as the alias rather than the reverse. As written, `eboot` is the canonical name in ebuild and `eBoot` is the exception — which is backwards from what GitHub says, and means the alias list has to keep growing rather than shrinking. Not worth reworking this PR over; worth knowing when the next one is added.

## Verification

Merged onto `origin/master` in sequence after #66, #67, #68, #69 and #73: **247 tests pass**. Not verified against `master` alone, which cannot collect its test suite — #66 repairs that and must land first.

PR #78, opened 2026-08-30
fix(ebuild): repair SyntaxError in dispatch.py and unify backend errors

prakhar7017:
## Summary

Fix a syntax error in `ebuild/build/dispatch.py` caused by duplicate `else:` clauses in `BackendDispatcher.configure()`.

The merge conflict left two different unknown-backend guards in the same `if` chain, causing `ebuild build`, `ebuild configure`, and `ebuild clean` to fail during import.

## Changes

- Removed the duplicated `else:` and consolidated the backend validation logic.
- Added `UnknownBackendError`, inheriting from both `ValueError` and `RuntimeError`, to preserve the existing caller contracts.
- Added explicit backend lists for each dispatcher operation:
  - `CONFIGURE_BACKENDS`
  - `BUILD_BACKENDS`
  - `CLEAN_BACKENDS`
- Removed `ninja` from the configure no-op path so unsupported `ninja` configuration fails with an actionable error.
- Kept `ninja` supported for `clean`, where no toolchain is required.
- Added six regression tests covering:
  - Dual `ValueError`/`RuntimeError` inheritance
  - `ninja` rejection during configure
  - Actionable error message
  - Correct backend lists in error messages
  - `clean("ninja")` continuing to work
  - Existing `cargo`, `make`, and `kbuild` no-op behavior

### Targeted tests

```text
pytest tests/ebuild/test_dispatch.py tests/unit/test_dispatch.py
31 passed

### CODE REVIEW

**Review by srpatcha — COMMENTED** [review]
Correct diagnosis and a correct fix — you and #66 found the same thing and
resolved it the same way.

I verified both repair `master`:

```
origin/master   SyntaxError: invalid syntax, line 133
#66             parses OK
#78 (this)      parses OK
#75             SyntaxError — still line 133, despite its title
```

Tracked as #87.

You also both reached the same conclusion about *why* the two error paths
existed. #66 puts it this way:

> the two blocks also disagreed about the exception — `tests/ebuild` expects
> `ValueError("Unknown build backend '<name>'")` while `tests/unit` expects
> `RuntimeError` matching "ninja". Both are legitimate readings: an unrecognized
> name is a bad argument, and a "ninja" that reaches the dispatcher is a CLI
> routing failure.

and resolves it with `UnknownBackendError(ValueError, RuntimeError)` so both
suites keep passing. "Unify backend errors" in your title is the same idea.

#66 is already approved and additionally restores
`NinjaBackend._object_path()`, which a cflags merge had deleted, so it is the one
that will land.

## What that leaves you

Nothing here is wrong; it is simply second to arrive at the same place. Rather
than resolving this against #66, I would close it and pick up something that is
not already covered — #85 (eos and eBoot defining colliding CMake target names,
which stops `ebuild integration` configuring) is open and in this area.

If any part of this diff does something #66 does not, say which and I will look
again — I compared the `dispatch.py` outcomes and the error-type reasoning, not
every line of the tests.

Two people independently identifying the same splice and choosing the same
resolution is a good signal about the fix. It is also a signal about the process
that let it land: `required_status_checks` is `null` on this repository, so
nothing imports the merge result before it becomes `master`. That is the part
worth fixing once #66 is in.

PR #79, opened 2026-08-30
Fix/ninja object path missing

prakhar7017:
## Summary

Fix a syntax error in `ebuild/build/dispatch.py` caused by duplicate `else:` clauses in `BackendDispatcher.configure()`.

> **Stacked on #78** (`fix/dispatch-syntax-error-merge-artifact`). Merge that first — without it `ebuild/build/dispatch.py` does not parse and the suite cannot be collected.

## Problem

`NinjaBackend.generate()` cannot produce a build. Both writers call
`self._object_path(...)`, but the method does not exist:

The merge conflict left two different unknown-backend guards in the same `if` chain, causing `ebuild build`, `ebuild configure`, and `ebuild clean` to fail during import.

## Changes

- Removed the duplicated `else:` and consolidated the backend validation logic.
- Added `UnknownBackendError`, inheriting from both `ValueError` and `RuntimeError`, to preserve the existing caller contracts.
- Added explicit backend lists for each dispatcher operation:
  - `CONFIGURE_BACKENDS`
  - `BUILD_BACKENDS`
  - `CLEAN_BACKENDS`
- Removed `ninja` from the configure no-op path so unsupported `ninja` configuration fails with an actionable error.
- Kept `ninja` supported for `clean`, where no toolchain is required.
- Added six regression tests covering:
  - Dual `ValueError`/`RuntimeError` inheritance
  - `ninja` rejection during configure
  - Actionable error message
  - Correct backend lists in error messages
  - `clean("ninja")` continuing to work
  - Existing `cargo`, `make`, and `kbuild` no-op behavior

### Targeted tests

```text
pytest tests/ebuild/test_dispatch.py tests/unit/test_dispatch.py
31 passed

### CODE REVIEW

**Review by srpatcha — COMMENTED** [review]
Thanks for this — the diagnosis is right, but the fix has already landed by
another route, so I want to save you the rebase.

This restores `NinjaBackend._object_path()`, which a cflags merge had deleted. #66 restores it too, with target-namespaced object paths and the same worked example:

```python
>>> backend._object_path(target, "src/main.c")   # target.name == "app"
PosixPath('_build/obj/app/src/main.o')
```

and the same reasoning — two targets may legitimately list one source (a library
and a test binary sharing a helper, or one source built twice with different
defines), so keying only on the source made both claim one output and ninja
rejected it with "multiple rules generate ...".

## Where that leaves this PR

It also touches `dispatch.py`, which conflicts with #66 — already approved, and
the PR that repairs `master`'s SyntaxError (#87). Verified: this branch does not
apply cleanly on top of it.

Rather than resolving the conflict, I would close this. Nothing in it is wrong;
it is second to arrive at the same place.

If part of this does something #66 and `49cd368` do not, say which and I will
look again — I compared the object-path namespacing and the escaping semantics,
not every line.

## Something open, if you want it

#85 — `eos` and `eBoot` both define CMake targets named `test_crypto` and
`test_multicore`, and since `ebuild` composes them with `add_subdirectory`, the
integration build cannot configure at all. Same class of bug as the ones you have
been finding, one repository wider.

srpatcha (dismissed):

PR #82, opened 2026-08-30
fix: invalidate cache when install directory is missing

nikhilramakrishna3005:
## Summary

Fixes a stale package-cache state where `PackageCache.is_built()` could return `True` even when the cached package's `install/` directory no longer existed.

A valid `.built` marker alone does not guarantee that the cached package is usable, because callers rely on the `install/` directory for the package's installed headers and libraries.

## Type of Change

- [ ] feat — New feature
- [x] fix — Bug fix
- [x] test — Add or fix tests

## Changes

- Updated `PackageCache.is_built()` to require the package `install/` path to exist as a directory before treating the cache as valid.
- Added a regression test that verifies a valid `.built` marker is not sufficient when the corresponding `install/` directory has been removed.

## Testing

- [ ] Integration tests pass

Focused regression test:

`python3 -m pytest tests/ebuild/test_package_cache.py -v --tb=short`

Result: **1 passed**

Lint:

`python3 -m flake8 ebuild/packages/cache.py tests/ebuild/test_package_cache.py`

Result: **passed**

The regression test was also verified against the original implementation, where it failed because `is_built()` incorrectly returned `True` after the `install/` directory was removed.

The broader test suite currently has unrelated failures on upstream `master`, including existing dispatcher and Ninja backend issues. The new package-cache regression test passes independently.

## Pre-Submission Checklist

- [ ] All existing tests pass
- [x] Documentation updated if API changed

## Related Issues

None.

## Screenshots / Logs

Not applicable.

## Additional Notes

This change uses `is_dir()` rather than only `exists()` because a non-directory path named `install` is not a usable package installation prefix.

The existing `.built` metadata validation for package name, version, and checksum is unchanged. No public APIs or cache metadata formats are modified.


srpatcha (approved):
Approving. Small, correct, and the test genuinely pins the behaviour.

`is_built()` is read by callers as "the cached install prefix is usable", but it
was only checking the `.built` marker. A marker outliving its install directory —
a partially cleaned cache, an interrupted `rm`, a pruned build tree, anything
that removes `install/` without removing the marker beside it — made
`is_built()` return true for a prefix that is not there. The build then skips
the rebuild and fails later, somewhere that gives no hint the cache was the
cause.

Checking both is the right shape: the marker says "we finished", the directory
says "and the result still exists", and the caller needs both to be true.

## Verified

Applied onto #66 (necessary — `master` itself does not parse, #87), clean apply:

```
with fix       5 passed, 202 deselected
without fix    FAILED test_is_built_is_false_when_install_dir_is_missing
               1 failed, 4 passed
```

Removing only your two-line guard turns the test red, so it fails on the unfixed
code and passes on the fixed one. That is what separates a regression test from a
demonstration, and a surprising number of "regression tests" do not manage it —
one earlier in this repo asserted on values that were correct both before and
after the bug it claimed to cover.

Whole suite on the same base: **206 passed**, up from 205, no failures.

The docstring naming the invariant — "a valid `.built` marker is not enough if
`install/` is gone" — is the part that will stop someone deleting this check in
two years as a redundant stat call.

Needs #66 merged first for CI here to run at all.

PR #84, opened 2026-08-30
feat(cli): add ebuild test and ebuild monitor — the last two golden-path steps

### LINKED ISSUE #83: MVP golden path is missing 2 of its 8 commands: ebuild test and ebuild monitor
§7 names eBuild's MVP responsibility as:

> **eBuild** — Setup, configure, build, **test**, flash, **monitor**.

§7.1 gives the golden path verbatim:

```
ebuild setup
ebuild new hello-eos
cd hello-eos
ebuild configure --board <reference-board>
ebuild build
ebuild test
ebuild flash
ebuild monitor
```

> Success criterion: a new developer should reach a verified first firmware run
> quickly and repeatably.

Two of the eight commands do not exist:

```
ebuild setup      EXISTS
ebuild new        EXISTS
ebuild configure  EXISTS
ebuild build      EXISTS
ebuild test       MISSING
ebuild flash      EXISTS
ebuild monitor    MISSING
```

Full command list today:

```
add analyze build clean configure firmware flash generate-board generate-boot
generate-project info install integration list-packages models new package
pipeline qemu repos sdk setup system
```

Twenty-three commands, and neither of the two the MVP definition names.

## What this costs

The golden path stops at step 5 of 8. `test` is the step that makes the run
*verified* rather than merely completed, which is the stated success criterion —
without it a developer builds, flashes, and then has no way to find out whether
what they flashed works. `monitor` is how they would see it run at all.

`ebuild qemu` exists and is adjacent to `monitor`, but it is a different thing:
it launches an emulator, not a serial console onto whatever the developer just
flashed. Step 7 flashes real hardware; step 8 has to be able to watch real
hardware.

## Scope

Two commands, following the CLI conventions already in `ebuild/cli/commands.py`:

- **`ebuild test`** — build the project's tests and run them, reporting the
  counts the underlying runner printed rather than inferring them from exit
  status. Must fail when tests fail; a test command that cannot go red is worse
  than none.
- **`ebuild monitor`** — open a serial console on the target's port, with the
  baud rate from the board configuration, `--port` and `--baud` overrides, and a
  clean exit on Ctrl-C. Must report an actionable error when no device is
  present rather than hanging.

Both need to work without hardware attached to be testable in CI: `test` runs on
the host, and `monitor` must fail cleanly and immediately with no port.

srpatcha:
Closes #83.

§7 names eBuild's MVP responsibility as "Setup, configure, build, **test**,
flash, **monitor**", and §7.1 gives the sequence verbatim. Two of the eight
commands did not exist:

```
ebuild setup      EXISTS        ebuild build      EXISTS
ebuild new        EXISTS        ebuild test       MISSING
ebuild configure  EXISTS        ebuild flash      EXISTS
                                ebuild monitor    MISSING
```

Twenty-three commands were present, and neither of the two the MVP definition
names. The golden path stopped at step 5 of 8.

`test` is the step that makes a first firmware run *verified* rather than merely
completed — the stated success criterion. `monitor` is how a developer sees it
run at all. `ebuild qemu` is adjacent but different: it starts an emulator and
cannot observe the board step 7 just flashed.

## `ebuild test`

Dispatches through the existing backend detection and reports what the runner
printed. Counts are parsed from the runner's own summary for ctest, meson and
cargo, and **withheld** when nothing recognisable was printed — a number derived
from a zero exit status is a guess presented as a measurement. `make` has no
standard summary, so its counts stay unknown and the exit status carries the
verdict.

The case that drove the design: **ctest exits 0 when it finds nothing to run.**
Trusting the exit status reports a green suite for a project containing no tests
at all — worse than having no test command, because it actively reassures. An
empty run is therefore treated as a failure.

Verified against all three outcomes on a real CMake project:

```
zero tests      [error] The runner completed without executing a
                single test.                                   exit=1

1 pass 1 fail   [error] Tests failed (1 passed, 1 failed).      exit=1

1 pass 0 fail   [ok] Tests passed (1 passed, 0 failed).         exit=0
```

## `ebuild monitor`

Opens a serial console, reading the baud rate from `board.console.baud` or
`monitor.baud` in `build.yaml`, falling back to 115200.

It auto-selects a port only when exactly one is present; with several it lists
them and asks, because picking one at random is how a session ends up silently
attached to the wrong board. An explicit `--port` is always honoured, since
enumeration misses some adapters. `--list` shows what the host can see. Ctrl-C
ends the session and closes the port — leaving it held makes the next run fail
with a busy device that looks like a hardware fault.

## A .gitignore trap, fixed in passing

Hit while adding these. The rule `build/` matched `ebuild/build/` — the backend
package, not build output:

```
$ touch ebuild/build/probe.py
$ git status --porcelain ebuild/build/probe.py
(nothing)
```

A new module added there is silently untracked: it works locally and is simply
absent from the repo. All four current files predate the rule so nothing has
been lost yet, but it would have caught this PR had `monitor.py` gone under
`ebuild/build/` rather than `ebuild/firmware/`.

Fixed with a negation rather than re-anchoring to `/build/`, so
`examples/*/build/` stays ignored. Both directions verified:

```
ebuild/build/probe.py       ??  (now visible)
examples/*/build/out.o      still ignored
```

## Verified

```
tests    99 -> 121 passed, 0 failed
```

The 22 new tests cover the count parsers for all four backends, the empty-run
detection by both routes (marker phrase and zero totals), and every branch of
port resolution and baud resolution — including that a malformed `build.yaml`
falls back rather than raising.

README updated with the full golden path and the two commands' behaviour.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,12 @@ jobs:
         run: ruff check . --select=E,F,W --ignore=E501
         continue-on-error: true
 
+      # --exclude: layers/eosuite/ vendors its own tests/ package, so a bare
+      # `mypy .` sees two modules named "tests" and bails with "Duplicate
+      # module named 'tests'" before checking anything. continue-on-error hid
+      # that the type check was doing no work at all.
       - name: Type check (mypy)
-        run: mypy . --ignore-missing-imports --no-strict-optional
+        run: mypy . --ignore-missing-imports --no-strict-optional --exclude '^layers/'
         continue-on-error: true
 
       # Runs the whole tests/ tree. The previous steps ran only tests/unit/
@@ -57,7 +61,13 @@ jobs:
       # coverage policy already lives in codecov.yml; whether to also
       # enforce it here is a maintainer decision, so this change leaves
       # both of those numbers alone.
+      #
+      # shell: bash — the matrix includes windows-2022, where the default
+      # shell is PowerShell and the backslash line continuations below are a
+      # syntax error ("Missing expression after unary operator '--'"), so the
+      # Windows leg of this job failed before pytest ever started.
       - name: Run test suite
+        shell: bash
         run: |
           python -m pytest tests/ -v --tb=short \
             --cov=ebuild --cov-report=xml --cov-report=term-missing \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,15 @@ on:
   pull_request:
     branches: [master, main]
 
+# Every other workflow in this repo declares a concurrency group; ci.yml,
+# the heaviest one, did not. Pushing twice to a PR left the earlier run
+# queued, and both competed for the same scarce windows/macos runners --
+# three superseded runs sat ahead of the current one for over an hour.
+# cancel-in-progress because a superseded commit's result is not wanted.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   # ── Test Matrix ───────────────────────────────────────────────────────────
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,12 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12"]
-        os: [ubuntu-22.04, macos-13, windows-2022]
+        # macos-latest, not macos-13: the macos-13 image is retired, so jobs
+        # requesting it queue forever and never get a runner. Across four runs
+        # on this branch the ubuntu-22.04 and windows-2022 jobs all started and
+        # finished while the three macos-13 jobs sat queued for over two hours.
+        # Every other workflow in this repo already uses macos-latest.
+        os: [ubuntu-22.04, macos-latest, windows-2022]
       fail-fast: false
 
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,10 @@ __pycache__/
 *.egg
 dist/
 build/
+# ...but ebuild/build/ is the backend package, not build output. Without
+# this a new module added there is silently untracked: git status shows
+# nothing at all, so it works locally and is simply absent from the repo.
+!ebuild/build/
 .eggs/
 *.whl
 
--- a/README.md
+++ b/README.md
@@ -32,8 +32,11 @@ Observed in the source tree:
   NuttX.
 - **Package graph** — install and list project packages
   (`ebuild add`, `ebuild list-packages`).
-- **Firmware & flashing** — build RTOS firmware and flash images to targets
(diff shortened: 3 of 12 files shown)


srpatcha:
Correction to this PR, and it is one worth stating plainly: **the numbers I
originally posted were measured against the wrong base.**

This branch was cut from a stale local `master` and was **75 commits behind**
`origin/master`. The "121 → 132 passed" figures described a tree that is not what
this would merge into. That is the same "green branch, base has moved, nothing
re-verified the result" failure I have been filing against other PRs here, and I
did it myself.

## Rebased and re-measured

Rebased onto #66 rather than `origin/master`, because `origin/master`'s
`ebuild/build/dispatch.py` does not parse:

```
$ python3 -c "import ast; ast.parse(open('dispatch.py').read())"
  line 133
        else:
        ^^^^
SyntaxError: invalid syntax
```

An orphan `else:` after a `raise` block — another merge splice. #66 repairs it
and is already approved, so this now stacks on it. Merge #66 first.

Corrected figures, measured against this branch's actual base:

```
base (#66)      205 passed, 1 skipped
this branch     240 passed, 1 skipped
new failures    none
```

The golden path re-verified after the rebase:

```
ebuild setup / new / configure / build / test / flash / monitor    all present
ebuild test on a real CMake project   ->  1 passed, 0 failed, exit 0
```

## And a defect I introduced during the rebase, since it makes the point

Resolving the `dispatch.py` conflict by keeping both sides spliced `test()` into
the middle of `clean()`, orphaned `clean()`'s body under a second `def clean`,
and dropped the module-level `TestOutcome` dataclass altogether:

```
ImportError: cannot import name 'TestOutcome' from 'ebuild.build.dispatch'
NameError: name 'dry_run' is not defined        (6 of master's own clean() tests)
```

Caught by running the suite and diffing the failure list against the base, not by
reading the diff. Fixed in the follow-up commit.

It is the identical shape to eos's `sync.c`, where #55 and #67 implemented
priority inheritance with incompatible designs and the squash spliced half of
each together. A conflict resolution is a code change and needs the same
verification as one.

srpatcha:
Superseded by #89.

`ebuild test` and `ebuild monitor` have both landed on `master` from another PR,
so the main body of this work is redundant — verified all eight golden-path steps
now resolve on a clean clone:

```
ebuild setup / new / configure / build / test / flash / monitor    all EXISTS
```

Two things here are **not** on master, and #89 carries them:

**The empty-run guard.** Master's `ebuild test` reports `[ok] All tests passed.`
and exits 0 for a project with zero tests — ctest exits 0 on "No tests were
found!!!", and a `CMakeLists.txt` with `enable_testing()` and no `add_test()`
still produces a `CTestTestfile.cmake`, so the runner is found and the wrapper
trusts the status. Confirmed against current master.

**The docs check** — that every `ebuild` command shown in the docs exists, code
blocks only, verified to fail when a fake command is added.

Closing rather than rebasing: the overlap is large enough that a narrow PR
against master reviews better than a conflict resolution against a branch that
has been overtaken.

PR #100, opened 2026-09-01
fix(cli): repair a merge splice that left commands.py unparseable

Kartikey1306:
## master does not parse

`ebuild` cannot start, and pytest collects nothing:

```
$ python -c "import ebuild.cli.commands"
  File "ebuild/cli/commands.py", line 855
    """ebuild — A unified embedded OS build system."""
              ^
SyntaxError: invalid character '—' (U+2014)
```

All 9 `Test` jobs on master fail this way, on every OS and Python version.

## Line 855 is not the problem

It is an ordinary docstring. The damage is 300 lines earlier — `_board_config()` lost its body **and its closing quotes** to a copy of `_resolve_backend_request()`'s last line, spliced in during a conflict resolution:

```python
def _board_config() -> Optional[Dict[str, Any]]:
    """The project's own board description, if it ships one.

    return resolved_backend, backend_config


def _configure_ninja_backend(
```

Its docstring never closes, so from there on every triple-quote toggles the wrong way, and the first em-dash the tokenizer reaches outside a string gets reported — at line 855, inside a function that is perfectly fine.

`_resolve_backend_request` still has its own `return` at line 364, so that line is a stray duplicate, not a move.

Introduced by `6d2c149`, a "Merge branch 'master' into …" commit, and carried to master through `a9e3c60` and `562d28d`.

## The same resolution dropped #89 entirely

`_NO_TESTS_MARKERS`, `_TEST_COUNT_PATTERNS`, `_ran_no_tests`, `_parse_test_counts` **and both call sites** are gone — the tests were left behind. `test` is back to:

```python
    log.success("All tests passed.")
```

reached unconditionally on a zero exit. ctest exits 0 when it finds nothing to run, so a project with `enable_testing()` and no `add_test()` reports a pass having executed nothing — exactly what #89 fixed.

## This PR

Restores both from `f5209ac`, the last commit where this file parsed. Verbatim; no new design.

```
$ python -c "from ebuild.cli.commands import _board_config, _ran_no_tests, _parse_test_counts; \
             print(_board_config(), _ran_no_tests('ctest','No tests were found!!!'), \
                   _parse_test_counts('ctest','96% tests passed, 1 tests failed out of 28'))"
None True (27, 1)
```

| | before | after |
|---|---|---|
| `import ebuild.cli.commands` | SyntaxError | ok |
| passing | 0 (nothing runs) | **474** |

## Still failing: 51

Pre-existing, and invisible until the module parses again. Not one cause — the largest groups:

- **9** — `NameError: name '_ninja_path' is not defined`. The function was renamed `escape_ninja_path` in `ninja_backend.py` and the tests were not updated (same shape as the `version_key` → `version_sort_key` rename in #98). Happy to take this in a follow-up.
- **10** — `dispatcher.configure(...) ran, but build.yaml declares targets` — the declared-targets precedence guard is not in effect.
- **8** — `--build-dir` resolution.
- **7** — `test_dispatch`.
- the rest spread across `doctor`, `package`, footprint and config validation.

> This is the fourth time implementation hunks have been lost to conflict resolution here (#64, #72, #77, #89, and now this). The merges are not gated: `f5209ac` and `562d28d` both reached master with no CI run on the merge commit. Requiring `CI — ebuild` on master before merge would have caught every one of them.

### DIFF
--- a/ebuild/cli/commands.py
+++ b/ebuild/cli/commands.py
@@ -521,7 +521,18 @@ def _report_footprint(cfg: "ProjectConfig", build_path: Path, log: Logger) -> No
 def _board_config() -> Optional[Dict[str, Any]]:
     """The project's own board description, if it ships one.
 
-    return resolved_backend, backend_config
+    A project that states its part's real capacity should not be measured
+    against the reference part for its family.
+    """
+    path = Path("board.yaml")
+    if not path.is_file():
+        return None
+    try:
+        import yaml
+        data = yaml.safe_load(path.read_text(encoding="utf-8"))
+    except Exception:
+        return None
+    return data if isinstance(data, dict) else None
 
 
 def _configure_ninja_backend(
@@ -2411,7 +2422,10 @@ def test(log: Logger, config_path: str, build_dir: str,
     log.info(" ".join(argv))
 
     try:
-        result = subprocess.run(argv, cwd=str(cwd))
+        # Captured rather than inherited, because the exit status alone cannot
+        # distinguish "every test passed" from "there were no tests". The
+        # output is echoed below so the terminal reads as it did before.
+        result = subprocess.run(argv, cwd=str(cwd), capture_output=True, text=True)
     except FileNotFoundError:
         log.error(
             f"{name} is not installed or not on PATH, so the tests cannot be "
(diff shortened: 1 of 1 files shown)


Kartikey1306:
Merge-order note: this restores collection, but master stays red if it lands alone — the 51 failures it exposes were always there, just unreachable behind the SyntaxError. #101 and #102 clear them. Green at the tip: **#102**. Merging 100 → 101 → 102 in order is fine, master is just red in between.

Kartikey1306:
Closing in favour of **#102**, which contains this commit (`662aecf`) verbatim as the first of its five.

This PR cannot go green on its own, and that is a property of the stack rather than of the change. `master` arrives with a test suite that does not collect at all, and repairing the splice in `commands.py` only gets it as far as *running*:

| | pytest |
|---|---|
| `master` `562d28d` | **11 errors during collection** — suite never runs |
| this PR `662aecf` | 51 failed, 474 passed |
| #101 `fcc97e6` | 27 failed, 498 passed |
| **#102 `51a4566`** | **525 passed, 0 failed** |

Each step is a strict improvement, but branch protection needs a green head, and the remaining failures are only cleared by the last three commits. Giving this PR enough to pass would just reproduce #102.

Nothing here is lost — #102 carries this commit unchanged, and is green on all 9 `Test` jobs (ubuntu/macos/windows × 3.10/3.11/3.12), Build Python Package, CodeQL, and all 11 EoSim boards.

PR #101, opened 2026-09-01
fix(ninja): the backend crashed on every target, and emitted an undefined rule

Kartikey1306:
> Stacked on #100, which makes `commands.py` parse again. Review the last commit.

## Generating a build file fails outright

```python
>>> NinjaBackend(cfg, Path("build"), toolchain).generate()
NameError: name '_ninja_path' is not defined. Did you mean: 'ninja_path'?
```

`_ninja_path` was renamed `escape_ninja_path`, and three call sites were not updated — the **executable** edge (line 223), the **static_library** edge (241) and the **shared_library** edge (257). Between them that is every target type the backend emits, so `configure` / `build` on the ninja backend produce no `build.ninja` at all.

Those three were also **double-escaping**: `out`, `obj_files` and `dep_archives` are each already escaped where they are built (lines 191, 217, 221, 238). So the wrapper is not merely undefined, it is redundant — dropped rather than renamed.

## Even fixed, a shared_library was unbuildable

The edge names a rule the generated file never defines. Only `cc`, `link` and `ar_rule` are emitted:

```
build build/libmylib.dylib: link_shared build/obj/mylib/lib.o

$ ninja
ninja: error: unknown rule 'link_shared'
```

The comment at line 246 already assumed the rule existed ("the flag itself lives in the `link_shared` rule, so it must not be repeated here") — the rule was simply never written.

This adds it, carrying the shared-object flag so no edge repeats it, plus `_shared_flag()` to choose it: `-dynamiclib` on macOS, `-shared` on ELF. Emitting the ELF spelling on macOS is what the two tests in this area were failing on; both name `_shared_flag()` in their comments.

## End to end

```
$ ninja -f build/build.ninja
[1/2] CC build/obj/mylib/lib.o
[2/2] LINK_SHARED build/libmylib.dylib
$ file build/libmylib.dylib
Mach-O 64-bit dynamically linked shared library arm64
```

On master the same project never gets as far as a build file.

## Tests

`tests/unit/test_ninja_backend.py` used the pre-rename name in five assertions (its import line already had the new one); updated.

| | before | after |
|---|---|---|
| the two ninja modules | 13 failed, 3 passed | **16 passed** |
| whole suite | 474 passed | **498 passed** |

### DIFF
--- a/ebuild/build/ninja_backend.py
+++ b/ebuild/build/ninja_backend.py
@@ -31,6 +31,16 @@ _PIC_FLAGS = {"-fPIC", "-fpic", "-fPIE", "-fpie", "-fno-pic", "-fno-PIC",
               "-fno-pie", "-fno-PIE"}
 
 
+def _shared_flag() -> str:
+    """The flag that makes the compiler driver emit a shared object.
+
+    macOS links dynamic libraries with -dynamiclib; ELF platforms use
+    -shared. It lives on the rule rather than on each edge so a
+    shared_library target cannot be emitted without it.
+    """
+    return "-dynamiclib" if sys.platform == "darwin" else "-shared"
+
+
 def escape_ninja_path(path) -> str:
     """Escape a path for use inside a Ninja build statement.
 
@@ -179,6 +189,13 @@ class NinjaBackend:
             "  command = $ar rcs $out $in",
             "  description = AR $out",
             "",
+            # A shared_library edge names this rule. Without the rule the
+            # generated build.ninja referenced an undefined rule and ninja
+            # refused the whole file.
+            "rule link_shared",
+            f"  command = $cc {_shared_flag()} $ldflags $in -o $out $libs",
+            "  description = LINK_SHARED $out",
+            "",
         ]
 
         toolchain_ldflags = self._get_toolchain_ldflags()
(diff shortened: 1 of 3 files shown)


Kartikey1306:
Merge-order note: stacked on #100. Master is green once **#102** lands; this one alone still leaves the doctor/package/precedence failures visible.

Kartikey1306:
Closing in favour of **#102**, which contains both of this branch's commits (`662aecf`, `fcc97e6`) verbatim as the first two of its five.

This PR cannot go green on its own. `master` arrives with a suite that does not collect, and this branch repairs part of the way there but not all of it:

| | pytest |
|---|---|
| `master` `562d28d` | **11 errors during collection** — suite never runs |
| #100 `662aecf` | 51 failed, 474 passed |
| this PR `fcc97e6` | 27 failed, 498 passed |
| **#102 `51a4566`** | **525 passed, 0 failed** |

The 27 that remain here are the same class the last three commits fix — implementations discarded in conflict resolution while their tests landed. Adding them to this branch would just reproduce #102.

The ninja work in `fcc97e6` (the backend crashing on every target, and the undefined rule it emitted) is unchanged in #102, which is green on all 9 `Test` jobs across ubuntu/macos/windows, plus Build Python Package, CodeQL, and all 11 EoSim boards.

PR #117, opened 2026-09-04
fix(deps): stop DepsManager setters mutating the shared DEFAULT_CONFIG template

oscar-leung:
## Issue

`DepsManager.load_config()` merged `DEFAULT_CONFIG` into the loaded config by reference (`data[key] = val`, `data["repos"][rname] = rval`) and seeded first-run configs with a shallow `dict(DEFAULT_CONFIG)`. The nested per-repo dicts in every instance's config were therefore the same objects as the module-level template. Since `setup()`, `link()`, `set_url()` and `set_branch()` all mutate those dicts in place, one instance's edits silently rewrote the process-wide defaults that every later instance (and any freshly written `~/.ebuild/config.yaml`) inherited.

## Approach

Deep-copy the default values at each merge point in `load_config()`. No behavioral change for the normal single-instance flow, and persisted configs are unaffected.

## Testing / validation

- Static analysis of all `DEFAULT_CONFIG` consumers and every `self._config` mutation site.
- Added `test_setters_do_not_mutate_module_defaults` to `tests/ebuild/test_deps_manager.py` (reusing the existing `isolated_deps` fixture): asserts the template stays pristine after `set_branch`/`set_url` while the instance's own config keeps and persists the edits.
- The test was written for CI to run; it was not executed locally.

## Limitations

The fix guards the default-template aliasing only; a caller who mutates `manager.config` directly can still alias into the live instance config (pre-existing API surface, unchanged here).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ebuild/deps/manager.py
+++ b/ebuild/deps/manager.py
@@ -10,6 +10,7 @@ locations instead of each command re-cloning into temp directories.
 
 from __future__ import annotations
 
+import copy
 import os
 import subprocess
 import warnings
@@ -70,17 +71,20 @@ class DepsManager:
         if EBUILD_CONFIG_PATH.exists():
             with open(EBUILD_CONFIG_PATH, "r", encoding="utf-8") as f:
                 data = yaml.safe_load(f) or {}
-            # Merge missing keys from defaults
+            # Merge missing keys from defaults. Deep-copied: setup(),
+            # set_url(), set_branch() and link() mutate these nested dicts
+            # in place, and an aliased default would leak one instance's
+            # edits into DEFAULT_CONFIG for the rest of the process.
             for key, val in DEFAULT_CONFIG.items():
                 if key not in data:
-                    data[key] = val
+                    data[key] = copy.deepcopy(val)
                 elif key == "repos" and isinstance(val, dict):
                     for rname, rval in val.items():
                         if rname not in data["repos"]:
-                            data["repos"][rname] = rval
+                            data["repos"][rname] = copy.deepcopy(rval)
             return data
 
-        self._config = dict(DEFAULT_CONFIG)
+        self._config = copy.deepcopy(DEFAULT_CONFIG)
         self.save_config()
         return self._config
 
--- a/tests/ebuild/test_deps_manager.py
+++ b/tests/ebuild/test_deps_manager.py
@@ -7,6 +7,7 @@ from pathlib import Path
 
 import pytest
 
+from ebuild.deps import DEFAULT_CONFIG
 from ebuild.deps.manager import DepsManager, SIBLING_DIR_NAMES
 
 
@@ -76,3 +77,25 @@ def test_sibling_eboot_missing_returns_none(isolated_deps, monkeypatch):
     resolved = DepsManager().get_repo_path("eboot", project_dir=project)
 
     assert resolved is None
+
+
+def test_setters_do_not_mutate_module_defaults(isolated_deps):
+    """In-place config edits must not leak into the shared DEFAULT_CONFIG.
+
+    load_config() used to alias the nested default repo dicts instead of
+    copying them, so set_branch() on one DepsManager rewrote the module-level
+    template — and every later instance in the process inherited the edit as
+    its 'default'.
+    """
+    mgr = DepsManager()
+    mgr.set_branch("eos", "dev")
+    mgr.set_url("eboot", "https://example.invalid/eboot.git")
+
+    assert DEFAULT_CONFIG["repos"]["eos"]["branch"] == "master"
+    assert DEFAULT_CONFIG["repos"]["eboot"]["url"].endswith("/eBoot.git")
+
+    # A fresh instance starting from a clean config must not see them either.
+    fresh = DepsManager()
+    fresh._config = fresh.load_config()
+    assert fresh.config["repos"]["eos"]["branch"] == "dev"  # persisted for us
+    assert DEFAULT_CONFIG["repos"]["eos"]["branch"] == "master"  # not for all


srpatcha (commented):
## Review — ebuild#117 "fix(deps): stop DepsManager setters mutating the shared DEFAULT_CONFIG template"
head: f598aa0fd79bbe28de78dc414139632ce8b36c92  author: oscar-leung  ci: pending (no check runs in bundle; `mergeStateStatus: BLOCKED`)

**Verdict:** The bug is real and the fix is correct — I reproduced the aliasing on `origin/master` and confirmed the patched tree clears it. The change is minimal and behaviour-preserving. Three smaller things below; none of them block the merge.

Evidence that the fix does what it claims (both trees extracted with `git archive`, driver replicating the new test's assertions with the same monkeypatching as `isolated_deps`):

```
--- BASE (unpatched) ---   FAIL
    DEFAULT_CONFIG["repos"]["eos"]["branch"]  == 'dev'  (want 'master')
    DEFAULT_CONFIG["repos"]["eboot"]["url"]   == 'https://example.invalid/eboot.git'
--- PATCHED ---            PASS
```

So the new test is not decorative: it fails without the production change and passes with it.

### Findings
|---|---|---|---|---|
| 1 | Medium (P2) | `ebuild/deps/manager.py:82` | The rewritten merge assumes `data["repos"]` is a dict. A `~/.ebuild/config.yaml` containing a bare `repos:` (no children — the single most likely hand-edit) parses to `None`, and `rname not in data["repos"]` raises an unhandled `TypeError`, killing every `ebuild` command that constructs a `DepsManager`. Verified against the patched tree: `repos:` → `TypeError: argument of type 'NoneType' is not a container or iterable`; `repos: []` → `TypeError: list indices must be integers or slices, not str`; `repos: 5` → `TypeError: argument of type 'int' is not a container or iterable`. Pre-existing, not introduced here — but these are the lines the PR is rewriting, and the malformed-input path is one the brief requires be covered. | Guard once at the top of the merge: `if not isinstance(data.get("repos"), dict): data["repos"] = {}`, and let the existing per-repo loop refill it. One line, and it turns a traceback into a repaired config. |
| 2 | Low (P3) | `tests/ebuild/test_deps_manager.py:94-101` | The test asserts against process-global `DEFAULT_CONFIG` and never restores it. Today that is safe because the fix keeps it pristine. If the fix is ever reverted or regressed, this test does not merely fail — it leaves `DEFAULT_CONFIG["repos"]["eos"]["branch"] == 'dev'` and `...["eboot"]["url"] == 'https://example.invalid/eboot.git'` for the rest of the session (confirmed on the unpatched tree). `tests/unit/test_sources_are_importable.py:125-131` then reads those same values and runs `git ls-remote --heads <url> <branch>` against them, so a regression in `manager.py` would surface as a confusing network failure in an unrelated file. | Add `monkeypatch.setattr("ebuild.deps.DEFAULT_CONFIG", copy.deepcopy(DEFAULT_CONFIG))`-style isolation, or an `autouse` fixture that snapshots and restores the template. Cheap, and it keeps the failure local to the test that caused it. |
| 3 | Low (P3) | `tests/ebuild/test_deps_manager.py:88-89` | `fresh = DepsManager()` already runs `load_config()` in `__init__`; the following `fresh._config = fresh.load_config()` repeats it and reaches into a private attribute to do so. It does not weaken the assertion, but it reads as if construction did not load the config. | Delete the line — `DepsManager()` alone proves the point. |
| 4 | Low (P3) | `ebuild/deps/manager.py:318` (adjacent, **not** introduced by this PR) | `unlink()` does `self._config.get("repos", {}).get(repo_name, {}).pop("path", None)`. For an unregistered name that pops from a throwaway dict and still calls `save_config()`; `ebuild/cli/commands.py:2251` passes an unvalidated `repo_name` and then prints `Unlinked {repo_name} — will use cached clone.` So `ebuild repos unlink typo` reports success having done nothing. `link()`, `set_url()` and `set_branch()` have the mirror problem: `setdefault(repo_name, {})` happily creates config entries for names outside `KNOWN_REPOS`. | Out of scope here. Worth a separate issue: validate `repo_name` against `KNOWN_REPOS` in the four setters, or in the `repos` CLI group. |

### Architecture conformance
Conforms. Per master design §21, `ebuild` is Tier 1 - Foundation and this is its own dependency-manager internals — correct repo, correct layer. §5.1's "eBuild understands the complete graph but is not a runtime dependency" is untouched: the change adds one stdlib import (`copy`) and no cross-repo dependency, and nothing here runs on a device. §9.2's "reproducible lockfiles/manifests for production builds" is served rather than harmed — a process-global template that one `DepsManager` could silently rewrite for the next is precisely a reproducibility hazard, and the first-run `~/.ebuild/config.yaml` written after such a mutation would have carried the leaked value onto disk permanently.

No public signature, serialized format, or manifest field changes; the on-disk `config.yaml` schema is identical and existing files are read exactly as before, so there is no §23.2 compatibility impact to declare. The PR body's stated limitation — that a caller mutating `manager.config` directly can still alias into the live instance config — is accurate and is a property of the existing API, not something this change introduces.

### Proposed changes
Only finding 1 is worth folding into this PR, and it is two lines:

```python
             for key, val in DEFAULT_CONFIG.items():
                 if key not in data:
                     data[key] = copy.deepcopy(val)
                 elif key == "repos" and isinstance(val, dict):
+                    if not isinstance(data["repos"], dict):
+                        data["repos"] = {}
                     for rname, rval in val.items():
                         if rname not in data["repos"]:
                             data["repos"][rname] = copy.deepcopy(rval)
```

with a test case alongside the new one asserting that a config file containing `repos:` loads to the full defaults rather than raising. Findings 2 and 3 are one-line test edits; finding 4 belongs in its own issue.

### Not checked
- **`pytest` was NOT RUN.** It is not installed on this host and `python3 -m venv` fails here (no `ensurepip`), so the repo's real suite — `python -m pytest tests/` per `.github/workflows/ci.yml:85` — did not execute. What I did instead: extracted `origin/master` and the patched tree with `git archive`, then drove `DepsManager` through the exact sequence and assertions of `test_setters_do_not_mutate_module_defaults`, replicating the `isolated_deps` fixture's monkeypatching by hand. The PASS/FAIL evidence above is from that driver, not from pytest's report. I have not observed the test pass under pytest's own collection.
- **ruff and mypy were NOT RUN** (`.github/workflows/ci.yml:49` installs both). The diff adds one used stdlib import and one used test import, so I expect no new diagnostic, but that is inferred.
- **CI result: unknown.** `checks.txt` in the bundle is empty — the PR was opened minutes before this run and no check runs had reported. `mergeStateStatus` is `BLOCKED`, consistent with required checks still outstanding; I did not inspect branch protection.
- The local `/home/srpatcha/eos/ebuild` clone is 100 commits behind `origin/master` and has a dirty working tree, so the sync step correctly left it alone. Every file I read for this review came from `origin/master` via `git cat-file`/`git archive`, not from the stale working copy.
- I did not audit the `~/.ebuild/config.yaml` upgrade path for configs written by older ebuild versions beyond the malformed cases listed in finding 1.


---
*Automated architecture review of `f598aa0fd79b` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

PR #124, opened 2026-09-10
fix(packages): add PackageRecipe serialization

OshanYelena:
## Summary

Fix package index synchronization failing with `AttributeError` when caching validated package recipes.

`IndexSyncManager.sync()` calls `PackageRecipe.to_dict()`, but `PackageRecipe` did not implement that method. This PR adds canonical recipe serialization using the existing YAML schema and adds regression coverage.

## Type of Change

- [ ] eat — New feature
- [x] fix — Bug fix
- [ ] docs — Documentation only
- [ ] style — Formatting, no code change
- [ ] 
efactor — Code restructuring without behavior change
- [x] test — Add or fix tests
- [ ] uild — Build system or dependency changes
- [ ] ci — CI/CD pipeline changes
- [ ] perf — Performance improvement

## Changes

- Added `PackageRecipe.to_dict()` to serialize package recipes using the canonical YAML schema.
- Added a regression test verifying that serialization uses external YAML keys such as `package` and `build` rather than internal model fields such as `name` and `build_system`.

## Testing

- [x] Unit tests pass
- [ ] Integration tests pass
- [ ] Manual testing performed
- [x] New tests added for new functionality

Tested with:

```bash
python -m pytest tests/unit/test_recipe.py tests/unit/test_index_sync.py -v
python -m pytest
```

## Pre-Submission Checklist

- [x] Code compiles without warnings (-Wall -Wextra -Werror for C)
- [x] All existing tests pass
- [ ] New tests added for new functionality
- [ ] Documentation updated if API changed
- [x] Commit messages follow <type>(<scope>): <description> convention
- [x] Branch is rebased on latest master

PR #125, opened 2026-09-10
fix(ebuild): preserve toolchain linker flags for shared libraries

### LINKED ISSUE #139: bug: Ninja shared-library links omit toolchain linker flags and sysroot
### What happened?

The Ninja backend omits toolchain linker flags and sysroot when linking shared-library targets. Executable and test targets include these settings, but shared libraries do not.

This can silently ignore linker checks such as -Wl,--no-undefined and omit the configured --sysroot.

The fix and regression tests are submitted in PR #125:
https://github.com/embeddedos-org/ebuild/pull/125

### Expected behavior

Shared-library links should include toolchain linker flags and sysroot before target-specific linker flags, matching executable targets. Neither the toolchain nor target flag lists should be mutated.

### Reproduction steps

1. Use the code before the fix in PR #125.
2. Configure a shared_library target with toolchain linker flags containing -Wl,--no-undefined and a sysroot.
3. Generate the Ninja build manifest.
4. Inspect the shared-library ldflags: the toolchain linker flag and --sysroot are missing.
5. On Linux/GCC, build a shared library containing an unresolved external symbol. The requested --no-undefined check is not enforced.

PR #125 adds manifest tests and a Linux/GCC CLI regression test for this behaviour.

### Version or commit

8b623d5786b5f841823f92a5506269f1e155a0f1 (baseline before the fix in PR #125)

### Environment

Linux, Python 3.12, Ninja backend, GCC.

Exact Linux distribution, GCC version and Ninja version were not recorded.

Real Windows/macOS linking and cross-compilation with an SDK were not tested.

### Logs or diagnostics

```shell
Historical verification recorded for PR #125:

pytest tests/unit/test_shared_library_toolchain.py -q

All three regression cases failed against the original backend and passed with the fix.

Full-suite result: 9 failed, 672 passed, 3 skipped.
The nine failures were pre-existing test_index_sync.py failures involving a missing PackageRecipe.to_dict method, also reproduced on the baseline.

These are recorded verification results, not a fresh test run.
```

### Submission checks

- [x] I searched existing issues for a duplicate.
- [x] This report contains no security-sensitive details.

SJ-14-SJ:
## Summary

Fix missing toolchain linker flags and sysroot when the Ninja backend links shared libraries. Shared-library targets now use the same computed toolchain settings as executable and test targets.

Implemented and tested with assistance from OpenAI Codex.

## Type of Change

- [x] fix — Bug fix
- [x] test — Add regression tests

## Changes

- Include `toolchain_ldflags` before target-specific linker flags in the shared-library branch of `ebuild/build/ninja_backend.py`.
- Add regression tests for linker flags, sysroot emission, flag ordering, and preservation of the original input lists.
- Add a Linux/GCC CLI regression test proving that `-Wl,--no-undefined` rejects an unresolved symbol.
- Add a CHANGELOG entry and retain the compact T-006 row in TASKS.md, as requested in review.

## Testing

The results below restore the original Linux/Python 3.12 verification record from commit `75d9249cebf639b986ec9032576cef70c515fdbd`, TASKS.md lines 39–66.

These are historical verification results, not a new test run during this description update. The backend code and regression tests are unchanged at commit `ed3a2eed4ff2d614f042db532700df1c217b5d4b`.

- [x] Regression tests added.
- [x] Focused regression tests passed in the original verification.
- [ ] Full repository suite passes — pre-existing failures documented below.
- [ ] Hosted CI successfully completed — not yet confirmed.

### Passing checks

- `pytest tests/unit/test_shared_library_toolchain.py -q`: all three regression cases failed against the original backend and passed with the fix.
- The original verification record reports that an independent reviewer repeated the focused tests: 3 passed.
- Ruff passed on both changed Python files.
- mypy passed on `ebuild/build/ninja_backend.py`.
- `python -m build --no-isolation` passed, producing an sdist and wheel.
- `git diff --check` passed.
- `yamllint .` passed; no YAML files were changed.

### Pre-existing failures

- Full pytest suite: 9 failed, 672 passed, 3 skipped in 51.88 seconds.
- All nine failures were in `tests/unit/test_index_sync.py`, reporting that `PackageRecipe` has no attribute `to_dict`.
- The same nine failures were reproduced in a detached baseline worktree at `8b623d5`: 9 failed and 13 passed in that module.
- Repository-wide mypy reported the same missing `PackageRecipe.to_dict` at `ebuild/packages/index_sync.py:354`.
- Repository-wide Ruff reported F811 in `tests/ebuild/test_build_dir_resolution.py`, W292 in `tests/ebuild/test_package_recipe.py`, and two E402 findings in `tests/unit/test_ci_gate.py`. These files are unchanged by this PR.

### Not run

- CMake/CTest: `cmake` was unavailable in the verification environment.
- Real Windows/macOS linker execution.
- Cross-compilation with an actual SDK/sysroot.

Portable manifest tests cover sysroot emission. The compiler-level regression is explicitly Linux/GCC-only.

## Related Issues

Fixes #139

## Additional Notes

The verification evidence is restored here in the PR description, while TASKS.md retains the compact task row requested in review.

No public API changes are introduced. Required CI checks and final maintainer review remain pending.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@
   `cjson` (v1.7.18), `nanopb` (v0.4.9.1), `lvgl` (v9.2.2), `tinyusb` (v0.18.0), and `unity` (v2.6.1).
 
 ### Fixed
+- Shared-library Ninja link commands now preserve toolchain `extra_ldflags`
+  and `sysroot`, before target linker flags and package library paths, matching
+  executable targets. Previously those toolchain settings were silently ignored.
 - **`ebuild test` now finds Windows test binaries.** The Ninja edge for a
   native `type: test` target already carried the platform suffix
   (`_exe_suffix()` names it `<name>.exe` on Windows), but `ebuild test`
--- a/TASKS.md
+++ b/TASKS.md
@@ -9,6 +9,62 @@ Status is one of: `todo`, `in-progress`, `blocked`, `review`, `done`.
 
 ## Active
 
+### T-119 — Preserve toolchain linker settings for shared libraries
+
+Owner: backend / testing
+Mode: Verification
+Status: review
+Depends on: none
+
+Goal: shared-library link commands honor the same toolchain settings as executables.
+
+Acceptance criteria:
+- Shared-library link flags include toolchain ldflags and sysroot before target ldflags and package library paths.
+- Generating multiple shared libraries does not mutate or leak flags between targets or into the toolchain.
+- A real Linux shared-library build honors toolchain `-Wl,--no-undefined`: a resolved source builds successfully and an unresolved source fails to link.
+
+Design: reuse the already-computed `_get_toolchain_ldflags()` result with list
+concatenation, matching executable handling. No API or dependency changes.
+Files in scope: `ebuild/build/ninja_backend.py`,
+`tests/unit/test_shared_library_toolchain.py`, `CHANGELOG.md`, `TASKS.md`.
+Other shared-library features and unrelated defects are out of scope.
+Risk: previously ignored linker options may now correctly reject invalid builds.
+Verification: new tests before/after the fix, full pytest suite, Ruff, mypy,
+Python package build, and the repository's CMake/CTest check.
+
+Handoff (planning/architecture -> implementation/testing): discovery found the
+shared-library branch initializes flags from the target alone. The executable
+branch already includes toolchain flags. Acceptance criteria are the three
+bullets above; all checks are NOT RUN. Next: write failing regressions, then fix.
+
+Verification results (Linux, Python 3.12):
+- PASS: all three new regression cases failed against the original backend,
+  then passed with the fix (`pytest tests/unit/test_shared_library_toolchain.py -q`).
+- PASS: independent reviewer repeated the focused tests (3 passed), found no
+  blocking issues, and confirmed all three acceptance criteria.
+- PASS: Ruff on both changed Python files; mypy on `ninja_backend.py`;
+  `python -m build --no-isolation` (sdist and wheel); `git diff --check`;
+  `yamllint .` (no YAML changes).
+- FAIL (pre-existing): full suite: `9 failed, 672 passed, 3 skipped in 51.88s`.
+  All nine failures are in `tests/unit/test_index_sync.py`:
+  `AttributeError: 'PackageRecipe' object has no attribute 'to_dict'`.
+  Reproduced the same nine failures in a detached baseline worktree at
+  `8b623d5` (9 failed, 13 passed in that module).
+- FAIL (pre-existing): repository-wide mypy reports the same missing
+  `PackageRecipe.to_dict` at `ebuild/packages/index_sync.py:354`.
+- FAIL (pre-existing): repository-wide Ruff reports F811 in
+  `tests/ebuild/test_build_dir_resolution.py`, W292 in
+  `tests/ebuild/test_package_recipe.py`, and two E402 findings in
+  `tests/unit/test_ci_gate.py`. These files are unchanged.
+- NOT RUN: CMake/CTest; `cmake: command not found` in this environment.
+- NOT RUN: real Windows/macOS linker execution and cross-compilation with an SDK.
+  Portable manifest tests cover sysroot emission; the compiler regression is
+  explicitly Linux/GCC-only.
+
+Handoff (verification -> maintainer): focused change independently reviewed;
+submission is for review, not a release. Remaining work: maintainer review/CI,
+with the unrelated baseline failures above tracked here for separate fixes.
+
 | ID | Task | Owner | Mode | Status | Depends on |
 |----|------|-------|------|--------|------------|
 | T-002 | Fix Windows Ninja test-target path parsing | backend | Maintenance | review | none |
--- a/ebuild/build/ninja_backend.py
+++ b/ebuild/build/ninja_backend.py
@@ -280,7 +280,7 @@ class NinjaBackend:
                     # get, which the rule preamble alone does not supply. The
                     # "build a shared object" flag itself lives in the
                     # link_shared rule, so it must not be repeated here.
-                    ldflags = list(target.ldflags)
+                    ldflags = toolchain_ldflags + list(target.ldflags)
                     libs = []
                     for pkg_name in target.uses:
                         pkg = self.package_paths.get(pkg_name)
--- a/tests/unit/test_shared_library_toolchain.py
+++ b/tests/unit/test_shared_library_toolchain.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026 EoS Project
+
+"""Regression tests for toolchain settings in shared-library link commands."""
+
+import shutil
+import sys
+from pathlib import Path
+
+import pytest
+from click.testing import CliRunner
+
+from ebuild.build.ninja_backend import NinjaBackend, PackagePaths
+from ebuild.build.toolchain import ResolvedToolchain
+from ebuild.cli.commands import cli
+from ebuild.core.config import ProjectConfig, TargetConfig
+
+
+@pytest.mark.parametrize("sysroot", [None, "/opt/target-sysroot"])
+def test_shared_link_preserves_toolchain_flags_without_mutation(tmp_path, sysroot):
+    toolchain = ResolvedToolchain(ldflags=["-Wl,--no-undefined"], sysroot=sysroot)
+    targets = [
+        TargetConfig(
+            name=name,
+            target_type="shared_library",
+            sources=[f"{name}.c"],
+            ldflags=[f"-Wl,-soname,lib{name}.so"],
+            uses=[name],
+        )
+        for name in ("first", "second")
+    ]
+    packages = {
+        name: PackagePaths(lib_dirs=[Path(f"vendor/{name}")], libraries=[name])
+        for name in ("first", "second")
+    }
+    config = ProjectConfig(name="libs", version="1.0", targets=targets)
+    NinjaBackend(config, tmp_path, toolchain, packages).generate()
+    manifest = (tmp_path / "build.ninja").read_text(encoding="utf-8")
+
+    for target in targets:
+        edge = next(
+            block for block in manifest.split("\n\n")
+            if ": link_shared " in block and f"lib{target.name}." in block
+        )
+        expected = ["-Wl,--no-undefined"]
+        if sysroot:
+            expected.append(f"--sysroot={sysroot}")
+        expected += target.ldflags + [f"-L{Path('vendor') / target.name}"]
+        assert edge.splitlines()[1:] == [
+            f"  ldflags = {' '.join(expected)}",
+            f"  libs = -l{target.name}",
+        ]
+        assert target.ldflags == [f"-Wl,-soname,lib{target.name}.so"]
+    assert toolchain.ldflags == ["-Wl,--no-undefined"]
+
+
+@pytest.mark.skipif(
+    not sys.platform.startswith("linux") or shutil.which("gcc") is None,
+    reason="requires Linux and GCC for --no-undefined shared-library linking",
+)
+def test_cli_shared_library_honors_toolchain_no_undefined(tmp_path, monkeypatch):
+    pytest.importorskip("ninja", reason="ninja Python package not installed")
+    monkeypatch.chdir(tmp_path)
+    (tmp_path / "build.yaml").write_text(
+        "project:\n  name: shared-link-check\n  version: '1.0'\n"
+        "toolchain:\n  compiler: gcc\n"
+        "  extra_ldflags: ['-Wl,--no-undefined']\n"
+        "targets:\n  - name: example\n    type: shared_library\n"
+        "    sources: [example.c]\n    cflags: [-fPIC]\n",
+        encoding="utf-8",
+    )
+    source = tmp_path / "example.c"
+    source.write_text("int example(void) { return 42; }\n", encoding="utf-8")
+    runner = CliRunner()
+    good = runner.invoke(cli, ["build", "--backend", "ninja", "--build-dir", "good"])
+    assert good.exit_code == 0, good.output
+    assert (tmp_path / "good" / "libexample.so").is_file()
+
+    source.write_text(
+        "extern int missing_symbol(void);\n"
+        "int example(void) { return missing_symbol(); }\n",
+        encoding="utf-8",
+    )
+    # A separate output directory avoids timestamp-based incremental decisions.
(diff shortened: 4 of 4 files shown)


srpatcha (commented):
## Review — ebuild#125 "fix(ebuild): preserve toolchain linker flags for shared libraries"
head: 75d9249cebf639b986ec9032576cef70c515fdbd  author: SJ-14-SJ  ci: none reported

**Verdict:** Clean. A one-line fix that makes the `shared_library` branch symmetric with the `executable`/`test` branch two blocks above it, a test that would have caught the original defect, and a CHANGELOG entry. I reproduced the generated manifest locally and it matches the test's expectations exactly, with no mutation of the toolchain or target flag lists. No new lint errors. The only things I would change are presentational.

### Findings
|---|---|---|---|---|
| 1 | Low | PR body | The body is the unmodified repository template: every Type-of-Change, Testing and Pre-Submission box unchecked, "Changes" left as two empty bullets, no linked issue. The actual verification record — which is thorough and unusually honest, including two `FAIL (pre-existing)` lines and two `NOT RUN` lines — is buried in `TASKS.md:51-77` where a GitHub reviewer will not look. `master` at 52e1f94 has just tightened the linked-issue policy. | Copy the `Verification results` block from `TASKS.md` into the PR body's Testing section and tick `fix` + `test`. The content already exists; it is in the wrong file for the audience. |
| 2 | Low | `TASKS.md:12-16`, `:23-77` | The T-119 entry is inserted above the `Active` table, and no `\| T-119 \| ... \|` row is added to that table — so the task does not appear in the ledger the file's header describes ("each owning role updates its own row"). It also uses inline `Goal:` / `Acceptance criteria:` rather than the definition-list form and the `\| Check \| Command \| Result \|` verification table that the T-000 template at `TASKS.md:65-96` specifies. | Add the `\| T-119 \| Preserve toolchain linker settings for shared libraries \| backend \| Verification \| review \| none \|` row to the Active table, and move the prose block below it in template form. The verification prose converts to the template's table almost line for line. |
| 3 | Low | `tests/unit/test_shared_library_toolchain.py:157-160` | The compiler-level regression — the test that actually proves `-Wl,--no-undefined` reaches the linker — is skipped on the macOS and Windows legs of the matrix. That is correct and the `reason=` string says so plainly, so this is not a hidden gap; noting it only so the coverage is not read as nine legs deep. On the three `ubuntu-22.04` legs it does run: `ninja>=1.11` is a hard dependency in `pyproject.toml:16`, so `pytest.importorskip("ninja")` will not skip after `pip install -e .`, and the runner has `gcc`. | None needed. Leave as is. |

### Architecture conformance
Conforms. §21 places `ebuild` in Tier 1 — Foundation; `ebuild/build/ninja_backend.py` is build-graph generation, which is squarely eBuild's job. §5.1 holds in both directions: the diff adds no import, no link line and no manifest entry, and nothing here becomes a runtime dependency — `eBuild understands the complete graph but is not a runtime dependency`.

The change is also the right *kind* of fix under §9.2, "One source of truth for CLI, VS Code and EoStudio": rather than re-deriving toolchain flags for the shared-library case, it reuses the already-computed `toolchain_ldflags` from `:219`, so executables, tests and shared libraries now all read the same value from `_get_toolchain_ldflags()` at `:137`. One source, three consumers. Per the brief's duplication check, this is the opposite of duplication — it removes an inconsistency rather than adding a second implementation.

Ordering is correct. `link_shared`'s command template at `:214` is `$cc -shared $ldflags $in -o $out $libs`, so toolchain `-L` paths and `--sysroot` precede the object files, and package `-l` entries land last in `$libs` where the linker needs them. Putting `toolchain_ldflags` ahead of `target.ldflags` matches `:236` exactly, which means a target can still override a toolchain flag by repeating it — the same precedence executables already had.

No API or wire-format change (brief item 8): `_write_ninja` is private, no signature moves, and the emitted `build.ninja` gains flags rather than changing shape. No hot path is touched (brief item 9) — this runs once per configure, over `self.config.targets`, with no new loop.

### Proposed changes
The code needs no change. Two housekeeping items, in order:

1. Fill in the PR body from the `TASKS.md` verification block, per finding 1.
2. Add the missing `T-119` row to the `Active` table and reshape the block to the T-000 template, per finding 2.

One observation for the maintainer rather than the author: this PR's `TASKS.md` block independently reports the same two pre-existing failures that #122 and #124 are each addressing — nine `tests/unit/test_index_sync.py` failures from the missing `PackageRecipe.to_dict`, and four repository-wide ruff findings. Three separate contributors have now diagnosed the same broken gate from scratch. That is the cost of leaving `master` red; landing #122 (or #132) and one of #119/#124 would stop it recurring.

### Not checked
- **pytest — NOT RUN.** pytest is not importable on this host, and the local `ebuild` clone has a dirty working tree, left untouched per the rules of engagement. Neither of the two new tests was executed by me. The author's claim that all three cases fail against the original backend and pass with the fix is unverified as a *pytest run*; I verified the underlying manifest assertion by other means, below.
- **The GCC link test — NOT RUN.** `test_cli_shared_library_honors_toolchain_no_undefined` was not executed. The claim that an unresolved symbol now fails the build is unverified.
- **mypy — NOT RUN.** Not installed. **yamllint — NOT RUN.** Not installed; no YAML in the diff.
- **CMake/CTest — NOT RUN**, matching the author's own `NOT RUN` line for the same reason.
- **CI — no checks reported** on `fix/shared-library-toolchain-ldflags`. GitHub returns no check runs for this branch at all, so nothing here has CI evidence behind it. `mergeable` and `mergeStateStatus` are both `UNKNOWN`.
- I did not test cross-compilation with a real sysroot; the `--sysroot=` emission is verified only as a string in the generated manifest.

**Verified locally:** against a `git archive` export of head `75d9249c` — (a) `ruff 0.16.5 check .` reports 4 errors, all of them the pre-existing `origin/master` findings, so this PR adds none; (b) driving `NinjaBackend(...).generate()` directly with `ResolvedToolchain(ldflags=["-Wl,--no-undefined"], sysroot="/opt/target-sysroot")` and two `shared_library` targets emits `ldflags = -Wl,--no-undefined --sysroot=/opt/target-sysroot -Wl,-soname,libfirst.so -Lvendor/first` and the matching line for `second`, and afterwards `toolchain.ldflags` and `target.ldflags` are both unchanged — the flag-leak acceptance criterion holds.


---
*Automated architecture review of `75d9249cebf6` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

srpatcha (dismissed):
Thanks — this is clearly right: `ninja_backend.py:283` now mirrors the executable branch at `:236`, and the `-Wl,--no-undefined` CLI test is a nice way to prove the flags reach the linker rather than only the manifest. I ran both new tests locally with `ninja` installed and they pass; the rest of the suite is unchanged, ruff and mypy are clean.

One request before merge:

- **`TASKS.md:12-66`** — the 56-line T-119 block sits ahead of the "Active" table and largely repeats the PR description (handoffs, local PASS/FAIL log). It will be stale the moment this merges. Could you reduce it to a table row in the existing format (or drop it, since the PR itself records the verification)? "T-119" also reads as a PR number; the next free task ID is T-006/T-007.
Optional: once #131 (automatic `-fPIC` for shared libraries) lands, the explicit `cflags: [-fPIC]` in `tests/unit/test_shared_library_toolchain.py:72` can be removed. Expect a trivial `CHANGELOG.md` conflict with #131/#126.

Approving with that nit.

srpatcha (commented):
## Review — ebuild#125 "fix(ebuild): preserve toolchain linker flags for shared libraries"
head: ed3a2eed4ff2d614f042db532700df1c217b5d4b  author: SJ-14-SJ  ci: fail

**Verdict:** Follow-up review. The code and tests are byte-identical to the head I reviewed at `75d9249c` — only `TASKS.md` and the PR body moved. Prior finding 2 is resolved; prior finding 1 is only half done, and the half that was skipped is now a red required check. One new finding: the verification record was deleted rather than relocated, so a behaviour-changing linker fix now carries no evidence anywhere.

### Status of the previous review's findings

| Prior # | Status | Evidence |
|---|---|---|
| 1 — PR body is the unmodified template | **Partially addressed** | A two-line summary was prepended to `pr.body`. Everything else is unchanged: all nine Type-of-Change boxes, all four Testing boxes and all six Pre-Submission boxes are still `- [ ]`; "Changes" is still `- ` / `- `; "Related Issues" is still empty. The recommended action — copy the verification block into Testing — did not happen; see finding 2 below for what happened to it instead. |
| 2 — `T-119` not in the Active table, wrong format | **Resolved in `ed3a2eed`** | `TASKS.md` now adds exactly one line: `\| T-006 \| Preserve toolchain linker settings for shared libraries ([#125](…)) \| backend \| Verification \| review \| none \|`, in the Active table, in the documented column order. The out-of-template prose block is gone. `T-006` does not collide with `T-002`–`T-005` in `origin/master`. |
| 3 — GCC link test skipped off Linux | **Untouched, and correctly so** | `tests/unit/test_shared_library_toolchain.py` hashes identical to the prior head (`b3be1864…`). The finding said no change was needed; none was made. |

### Findings
|---|---|---|---|---|
| 1 | High | CI — `policy / Policy / Linked Issue` | The required policy check **fails** (run 34816962349, 4s). `mergeStateStatus` is `BLOCKED` and `reviewDecision` is `REVIEW_REQUIRED`; `mergeable` is `MERGEABLE`, so this is the gate, not a conflict. The cause is visible in the PR body: the `## Related Issues` section is empty, which is the one part of prior finding 1 that was not addressed. This is the single thing standing between this PR and mergeability. | Open (or find) an issue for the shared-library toolchain-flag defect and put `Fixes #NN` under `## Related Issues`. Nothing in the tree needs to change. |
| 2 | Medium | `TASKS.md` (removed vs `75d9249c`) | The verification record was **deleted, not moved**. At `75d9249c`, `TASKS.md` carried a 30-line `Verification results (Linux, Python 3.12)` block — three `PASS` lines with commands, two `FAIL (pre-existing)` lines naming the nine `test_index_sync.py` failures and the four repo-wide ruff findings, and two `NOT RUN` lines for CMake/CTest and for Windows/macOS linking. That block is gone at `ed3a2eed` and nothing replaced it: not the PR body (all Testing boxes still unchecked), not the `Evidence` section, not the Done table. The `T-006` row says Mode `Verification`, Status `review`, and the ledger now holds no verification for it — which is exactly what `TASKS.md`'s own header says each owning role is responsible for. Per `.ai/reviewer.md` ("was anything deleted or overwritten without explanation?") and master design §28, the evidence for a behaviour-changing linker fix should not vanish in a formatting cleanup. | Restore the block, in the T-000 template's `\| Check \| Command \| Result \|` form, under the `Evidence` heading where `T-002`–`T-004` already keep theirs — or paste it into the PR body's Testing section, which is what prior finding 1 asked for. The text still exists verbatim at `git show 75d9249c:TASKS.md`, lines 39-66. |

No new defect in the new commits: there is no new code in them to defect. The one-line backend change and the 88-line test file are unchanged.

### Architecture conformance
Conforms; unchanged from the prior review and re-checked against the mirror rather than recalled.

§21 places `ebuild` in Tier 1 — Foundation, and `ebuild/build/ninja_backend.py` is build-graph generation, which is eBuild's own job. §5.1 holds in both directions: the diff adds no import, no link line and no manifest entry, and nothing here becomes a runtime dependency — "eBuild understands the complete graph but is not a runtime dependency". §9.2's "one source of truth" is what the fix actually restores: `toolchain_ldflags` from `_get_toolchain_ldflags()` now feeds executables, tests *and* shared libraries, instead of the shared-library branch re-deriving its own flag list. Per brief item 10 this removes duplication rather than adding it.

Brief item 8 (API/wire compatibility): none. `_write_ninja` is private, no signature changes, and `build.ninja` gains flags without changing shape. Brief item 9 (performance): none — this runs once per configure over `self.config.targets`, with no new loop or allocation. Brief item 4 (weakened checks): none; no test disabled, no lint loosened, no assertion removed.

### Proposed changes
1. Add `Fixes #NN` to `## Related Issues`. That clears the only red check and unblocks the PR.
2. While editing the body, tick `fix` and `test` under Type of Change, and fill the two empty `Changes` bullets — one line each for the backend change and the regression test.
3. Restore the deleted verification block under `TASKS.md`'s `Evidence` heading, in the T-000 table form.

None of these touch the code. The code is finished.

### Not checked
- **pytest — NOT RUN.** `pytest` is not importable on this host (`ModuleNotFoundError: No module named 'pytest'`), so neither new test was executed by me, at this head or any other. The author's claim that all three cases fail pre-fix and pass post-fix remains unverified as a pytest run — and as of this head that claim is no longer written down anywhere in the repo.
- **The GCC link regression — NOT RUN.** `test_cli_shared_library_honors_toolchain_no_undefined` was not executed. That an unresolved symbol now fails the link is unverified here.
- **mypy — NOT RUN** (not installed). **yamllint — NOT RUN** (not installed; no YAML in the diff). **CMake/CTest — NOT RUN.**
- **The commits between `75d9249c` and `ed3a2eed` — NOT INSPECTED as commits.** The local clone does not have object `ed3a2eed` (`git cat-file: could not get object info`), and the `ebuild` working tree is dirty (`TASKS.md`, `ebuild/cli/integration.py`, `tests/ebuild/test_integration_initramfs_security.py` modified), so per the rules of engagement I did not fetch into it. I compared *end states* instead, below; commit messages and any intermediate commit are unreviewed.
- Cross-compilation with a real sysroot is still untested; `--sysroot=` is verified only as a string in the generated manifest.

**Verified locally**, by exporting `origin/master` with `git archive` and applying the bundle's `diff.patch` to it:
- `git apply --check` then `git apply` — **applies cleanly**, confirming no conflict with `master`, consistent with `mergeable: MERGEABLE`.
- `sha256sum` of the resulting `ebuild/build/ninja_backend.py` (`dd109022…`) and `tests/unit/test_shared_library_toolchain.py` (`b3be1864…`) are **identical** to the same two files exported from `75d9249c`. This is the evidence for "code and tests unchanged since the last review", and it means the prior review's verification of the generated manifest still applies to this head.
- `ruff 0.16.5 check ebuild/build/ninja_backend.py tests/unit/test_shared_library_toolchain.py` — **All checks passed**, exit 0.
- `ruff 0.16.5 check .` on the patched tree — **4 errors**; the identical run on an unpatched `origin/master` export — **4 errors**. The PR introduces none; the four are the known pre-existing `F811`/`W292`/`E402` findings in unchanged test files.


---
*Automated architecture review of `ed3a2eed4ff2` — scheduled, model `claude-opus-5`, checked against the EmbeddedOS Master Design v2.0. Advisory only: this reviewer never approves, requests changes, or merges. Reply here to discuss or push back — a wrong finding is a bug worth reporting.*

